PR #1, opened 2017-06-12
Handle parsing Maven version ranges

kchodorow:
This isn't very smart and still won't cover specs that require
a remote lookup (e.g., (1.2.3,1.2.5)), but it eliminates all
warnings mentioned in https://github.com/bazelbuild/bazel/issues/1794
and https://github.com/bazelbuild/bazel/issues/3089.

Check out the test for examples of version ranges.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -24,6 +24,12 @@ maven_jar(
 )
 
 maven_jar(
+    name = "maven_artifact",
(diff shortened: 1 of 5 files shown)


kchodorow:
FYI @petroseskinder

katre (approved):

PR #4, opened 2017-06-12
Write output to generate_workspace.bzl by default, instead of WORKSPACE and BUILD files

kchodorow:
This splits the output writer into old & new versions. By default, `generate_workspace` will write generate_workspace.bzl, but users can specify `--direct-to-ws` to get the old behavior.

See BzlWriter.java#L54 for new usage info.


generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    outputStream.append("\n\n");
+
+    outputStream.println("def generated_java_libraries():");
+    if (rules.isEmpty()) {
+      outputStream.println("  pass\n");
+    }
+    for (Rule rule : rules) {
+      outputStream.println("  native.java_library(");
```
katre: Move this into a formatJavaLibrary method, similar to formatMavenJar.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/WorkspaceWriter.java
```diff
@@ -0,0 +1,128 @@
...
+  public void writeWorkspace(PrintStream outputStream, List<String> sources,
+      Collection<Rule> rules) {
+    writeHeader(outputStream, sources);
+    for (Rule rule : rules) {
+      outputStream.println(formatMavenJar(rule) + "\n");
+    }
+  }
+
+  @Override
+  public String formatMavenJar(Rule rule) {
```
katre: Should this move back to the AbstractWriter? I can't actually tell how it's different from the version in BzlWriter.

katre (commented):
Is it possible to add a shell-based integration test for this to ensure the generated rules work with Bazel? It'd also serve as an example of how to use the generator.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/WorkspaceWriter.java
kchodorow: It's prefixed by "native." and indented an extra 2 spaces :-/

katre: The logic is complicated enough I think it's still worth unifying, just passing in the initial indent and the rule name.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
kchodorow: Done.

kchodorow:
Wow, it is super annoying writing shell tests without unittest.bash. Working on that...

kchodorow:
That's a no on the shell tests. It's too hard to get a copy of Bazel (sadly) and generate_workspace contact Maven Central, which seems suboptimal.

I added more info to the README and Michelle's currently got a CL in progress documenting this.

katre (approved):

katre:
Too bad on the shell tests, but the README is definitely helpful.

Issue #25, opened 2017-06-14
Make generate_workspace visible


cpenner-va:
Can we set visibility to public on the generate_workspace? I'd love to be able to automatically pull the tool into some of my bazel projects so I can use it more easily, but the target is currently private.

petroseskinder:
Hi Chris, 

Thanks for the feedback. Can you explain your general workflow with `generate_workspace`? I saw your prior comments on #26 and #24, and I do acknowledge it is somewhat awkward that it cannot be called from within an existing project.

For this specific issue, I'll let @kchodorow answer whether this is an appropriate move. However, for the time being, you can modify `migration-tooling/generate_workspace/BUILD` file so that it reads:

```python
package(default_visibility = ["//visibility:public"]) #<-- new line to add

java_binary(
    name = "generate_workspace",
    main_class = "com.google.devtools.build.workspace.GenerateWorkspace",
    runtime_deps = ["//generate_workspace/src/main/java/com/google/devtools/build/workspace"],
)
```

cpenner-va:
Thanks; the temporary fix I was using was to just run directly from the external package, i.e. `bazel run @io_bazel_migration_tooling//generate_workspace`;

As far as I can tell the intended use-case for generate_workspace is that it should be used once when converting from one project type to another; we're using it every time that we need to add a new maven dependency.

The reason for this is that oftentimes maven dependencies have transitive dependencies and unifying and hunting them down is a large pain with a good chance of making mistakes.  Here's an example of how we're using it:

- Start new project
- Realize we need lib 'X', so generate a workspace with `generate_workspace -a com.X:X:1.0`
- Work some more...
- Realize we need lib 'Y' so generate a new workspace containing X AND Y via `generate_workspace -a com.Y:Y:1.0 -a com.X:Y:1.0`

In this way we know we never have duplicates and don't need to figure out transitive dependencies manually.

In order to ease this process we have a shell script which calls generate_workspace for us; something like:

```bash
#!/bin/bash
bazel run @io_bazel_migration_tooling//generate_workspace -- -a X:X:1.0 -a Y:Y:1.0 -o /project/dir
```

By including generate_workspace in our workspace or build file we guarantee that our build scripts and scripts to add libraries "just work" when other team members clone the repo. We don't need to ensure that they've installed the generate_workspace repo somewhere on their machine.

If you need more clarification or have any suggestions on how to improve our workflow then I'm very interested to hear them! Thanks for your work on this!

[...]

Issue #20, opened 2017-06-13
How to add a new maven deps and upgrade old maven deps to WORKSPACE file


petroseskinder:
# Migrated from [issue 989](https://github.com/bazelbuild/bazel/issues/989) in main repository
### Post 1
I see the generate WORKSPACE tool, which is useful.

But an ongoing issue is the pain in transitively adding the N^th dependency. Many of the transitive deps will already have been added, or added with different versions.

It is not clear the best way to tackle this problem.

Ideally, we could run a tool that would load the WORKSPACE, it would merge in a new POM and output a new updated WORKSPACE and BUILD. We would want to control what do to when we see redundant artifacts. Some ideas:

1. force the user to manually resolve.
2. just try the latest version and test.
3. use the tests to try to automatically pick a version (prefer later, but verify by running a test command).
4. set up a second, differently named external repo pointing at the jar. Find a way to error or warn if two versions of the same artifact wind up in the transitive deps of any one target.

Solution 4. might be ideal since actually putting exactly one version of every target might not be workable.

### Post 2
a related issue is upgrading deps. Many will need to be upgraded at a time. If we are using hashes in maven_jar those will all need to be updated and that could be a pretty tedious job by hand.

### Post 3
by the way, we have pretty much solved this problem with a tool we run along side bazel:
https://github.com/johnynek/bazel-deps

we maintain our maven dependencies in a yaml file, then it generates the transitive deps with shas and all the targets for the maven graph, normalized to one version per dependency.

It would be cool to have a repository rule do this from within bazel, but there is currently no way great way to do that which I can see, since we can't run jvm code from inside a repository rule. I guess we could pull a compiled jar, then run that jar to generate the dependencies, but so far our dependencies change rarely enough that an external tool is fine.

GinFungYJF:
For the Post 2, is there a good solution so far？

petroseskinder:
There will be after #46 is merged. Its a repository rule called `transitive_maven_jar`.

PR #29, opened 2017-06-15
Modified generate_workspace WORKSPACE file so that it uses the load.bzl format.

petroseskinder:
I was having an issue with adding a new dependency to a `maven_jar` and found it much easier adding a new dependency using the load.bzl format rather than using the old format.

I figure it makes more sense to refactor the WORKSPACE to this format, before presenting that issue. That said, something I found particularly awkward/cumbersome was that whenever I wanted to upgrade or add a new dependency, I would have to execute the following command:

```bash
java -jar generate_workspace_deploy.jar \
-a=com.google.code.findbugs:jsr305:3.0.1 \
-a=org.codehaus.plexus:plexus-utils:jar:3.0.24 \
-a=org.codehaus.plexus:plexus-component-annotations:1.7.1 \
-a=org.codehaus.plexus:plexus-interpolation:1.24 \
-a=com.google.guava:guava:20.0 \
-a=org.apache.maven:maven-artifact:3.5.0 \
-a=org.mockito:mockito-all:1.9.5 \
-a=junit:junit:4.4 \
-a=com.google.truth:truth:0.30
#<-- insert new thing here
```

Bah!😫 It would be nice if we didn't have to do all that. #20 and Chris's [comment](https://github.com/bazelbuild/migration-tooling/issues/25#issuecomment-308572362) in #25.[Stripe](https://github.com/johnynek/bazel-deps) uses an intermediary yaml file to make upgrading easier.

reviewer: @kchodorow 
cc: @hhclam


bazel-io:

generate_workspace/BUILD
```diff
@@ -1,3 +1,6 @@
+load('//:generate_workspace.bzl', 'generated_java_libraries')
+generated_java_libraries()
```
kchodorow: Remove.

BUILD
```diff
@@ -0,0 +1,2 @@
+load('//:generate_workspace.bzl', 'generated_java_libraries')
```
kchodorow: Remove (make it an empty file).

WORKSPACE
```diff
@@ -1,45 +1,2 @@
...
-    name = "org_junit",
-    artifact = "junit:junit:4.4",
-)
-
-maven_jar(
-    name = "com_google_truth",
-    artifact = "com.google.truth:truth:0.30",
-)
+load('//:generate_workspace.bzl', 'generated_maven_jars')
+generated_maven_jars()
```
kchodorow: Add newline.

kchodorow (commented):
Good idea. Just a few nits, the java libraries aren't actually used anywhere.

WORKSPACE
petroseskinder: Thanks. Are you using google's linter before you push code onto github? Any OSS alternatives? I really relied on it to avoid nits.

generate_workspace/BUILD
```diff
@@ -1,9 +0,0 @@
-load('//:generate_workspace.bzl', 'generated_java_libraries')
```
petroseskinder: Per your recommendations, I removed this file. However, with this removed, how are we now defining the java_binary?

petroseskinder: Removed the file, per your recommendations. However, with this removed, where is the java_binary defined?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/BUILD
```diff
@@ -18,7 +18,7 @@ java_library(
         "//third_party:maven_model",
         "@com_google_guava_guava//jar",
         "@com_google_code_findbugs_jsr305//jar",
-        "@org_apache_maven_maven_artifact//jar",
+        "@org_apache_maven_maven_artifgact//jar",
```
petroseskinder: nit

generate_workspace/BUILD
petroseskinder: Removed per recommendations. However, with this removed, where/when do we define the `java_binary`

petroseskinder:
> Good idea. Just a few nits, the java libraries aren't actually used anywhere.

That was a little confusing for me. Is the intended approach to use the java_library or depending directly the maven_jar? I can see how the java_library may be preferably, in that bazel won't try to fetch the java_library as it would the maven_jar.

generate_workspace/BUILD
```diff
@@ -1,6 +0,0 @@
-java_binary(
```
kchodorow: Don't delete this, just the load().

kchodorow (commented):
It would be fine to use the java_librarys (and we probably should switch to doing so at some point) but until we do, we don't need to load them.

generate_workspace/BUILD
petroseskinder: That makes significantly more sense.

kchodorow (approved):

PR #31, opened 2017-06-16
Fix jcommander artifact parsing so that it supports version ranges.

petroseskinder:
Previously, jcommander would parse `--a=group:artifact:[1.0,2.0]`
as {"group:artifact:[1.0", "2.0]"} rather than {"group:artifact:[1.0,2.0]"}.

jcommander defaults to splitting by comma, which we don't want.

reviewer:@kchodorow 
cc:@hhclam, @cgrushko, @jin


bazel-io:

kchodorow (approved):

PR #34, opened 2017-06-20
Modify version resolution s.t. we select the highest version possible 

### LINKED ISSUE #16: generate_workspace should/could choose highest maven version number
# Migrated from [issue 1573](https://github.com/bazelbuild/bazel/issues/1573) in main repository

If generate_workspace is run with artifacts that depend on a common artifact, it will choose the version from whichever it encounters first. Alternately, it would be better (I think) if it instead chose the highest version. For example:

```
% bazel run //src/tools/generate_workspace -- \
  --artifact=com.amazonaws:aws-java-sdk-core:jar:1.11.10 \
  --artifact=io.dropwizard:dropwizard-core:jar:1.0.0
```
Generates:
```
 # com.fasterxml.jackson.datatype:jackson-datatype-guava:bundle:2.7.6 wanted version 2.7.6
 # io.dropwizard:dropwizard-jackson:jar:1.0.0 wanted version 2.7.6
 # com.fasterxml.jackson.datatype:jackson-datatype-jsr310:bundle:2.7.6 wanted version 2.7.6
 # com.fasterxml.jackson.module:jackson-module-afterburner:bundle:2.7.6 wanted version 2.7.6
 # com.amazonaws:aws-java-sdk-core:jar:1.11.10
 # com.fasterxml.jackson.datatype:jackson-datatype-jdk8:bundle:2.7.6 wanted version 2.7.6
 maven_jar(
     name = "com_fasterxml_jackson_core_jackson_databind",
     artifact = "com.fasterxml.jackson.core:jackson-databind:2.6.6",
     sha1 = "5108dde6049374ba980b360e1ecff49847baba4a",
 )
```
But if we reverse the artifacts:

```
% bazel run //src/tools/generate_workspace -- \
  --artifact=io.dropwizard:dropwizard-core:jar:1.0.0 \
  --artifact=com.amazonaws:aws-java-sdk-core:jar:1.11.10
```
Now we get:
```
# com.fasterxml.jackson.datatype:jackson-datatype-guava:bundle:2.7.6
# com.fasterxml.jackson.module:jackson-module-afterburner:bundle:2.7.6
# com.amazonaws:aws-java-sdk-core:jar:1.11.10 wanted version 2.6.6
# io.dropwizard:dropwizard-jackson:jar:1.0.0
# com.fasterxml.jackson.datatype:jackson-datatype-joda:bundle:2.7.6
# com.fasterxml.jackson.datatype:jackson-datatype-jsr310:bundle:2.7.6
# com.fasterxml.jackson.datatype:jackson-datatype-jdk8:bundle:2.7.6
maven_jar(
    name = "com_fasterxml_jackson_core_jackson_databind",
    artifact = "com.fasterxml.jackson.core:jackson-databind:2.7.6",
    sha1 = "c7012a59b4f36843489753a2027e169c3b8586f8",
)
```
Choosing the highest would also ensure the output is the same regardless of which ordering is used for the artifacts.

For context, this is coming out of a desire to manage our dependencies with a simple list of maven artifacts, and then generating the `WORKSPACE` and `BUILD` files automatically. Currently, it seems like either we have to be careful with the ordering of our list of artifacts, or we have to modify the generated `WORKSPACE` after the fact.

petroseskinder:
Resolves #16.

### Context
When faced with multiple potential versions, Maven selects the highest matching version, if a range is provided. The resolver did not. Instead, it would guess a potential version, _without examining the available versions._ The following [design document](https://docs.google.com/document/d/12SgM5MzTAQJUmVg6HTQB7gb_sxT-OwxmRwsoySi34Sk/edit?usp=sharing) discusses the issues present with the prior implementation. In addition, it proposes a solution for how to incorporate backtracking and SAT solving to resolve versioning conflicts.

### What has changed?
* Commit 1: Resolver was refactored so that SHA1 downloading and versioning were in separate classes. Relevant tests were also moved to their own separate classes
* Commit 2: Version resolution was changed. Now, when requesting a version for an artifact, resolver scrapes maven central for all available versions, and selecting a version that fits within the provided range. Prior implementation had a number of edge cases that were not in compliance with how maven specs. Tests were added for each case.
    * For example, (2,3) would throw an exception. "2.3" did not get the highest version. [2.3,) would return 2.3, irrespective of whether there were later versions. (,4.0] would unintentionally throw an exception. Not providing a version would also throw an exception, instead of returning highest version.

### What is not implemented?
* **Support for Sequences of Versions**: For example, one can provide a version specification like "(,1.1),(1.1,). This excludes version 1.1. 
* **Support for Non-maven central repositories**: 
* **Version Conflict Resolution**: This is a naive greedy implementation. The emphasis is on resolving versions for a given artifact. It is unknown how it will behave on POM files or multiple artifacts with conflicting versions. However, it will perform no worse, than the existing implementation.

**Reviewers:** @kchodorow, @katre 
**CC:** @hhclam, @jin


bazel-io:

petroseskinder:
Sorry for the large CL. Much of the 600 lines comes from refactoring and writing test cases.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
```diff
@@ -1,47 +1,117 @@
...
+      // Already a version range.
+      return versionSpec;
+    }
+    return "[" + versionSpec + "," + ")";
+  }
+
+  /** Checks if a version is a snapshot versions, i.e. non-release versions */
+  // TODO add support for sequences of versions like "[2.0,2.4),(2.5,)"
+  private static boolean isSnapshotVersion(String version) {
+    return !Pattern.compile("\\d+(\\.\\d+)*(\\.\\d+)*").matcher(version).matches() || version.isEmpty();
```
petroseskinder: I chose not to implement support for version ranges like `(2.0,2.4),(2.5,)` The regular expression would be very gnarly.  The only alternative I could think of was defining a small grammar and using ANTLR to generate a parser, which seemed rather excessive. Thoughts?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ShaDownloader.java
```diff
@@ -0,0 +1,42 @@
...
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.util.logging.Logger;
+
+class ShaDownloader {
+
+  /**
+   * Downloads the SHA-1 for the given artifact.
+   */
+  static String downloadSha1(Rule rule, Logger logger) {
```
kchodorow, line 18: Create 1 logger per class, don't pass it around.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /** Checks if a given version is within the bounds of a version range */
+  private static boolean containedInSpecifiedRange(
+      String groupId, String artifactId, String versionSpec, String potentialVersion)
+      throws Resolver.InvalidArtifactCoordinateException {
+
+    if (versionSpec.isEmpty()) {
+      return true;
+    }
+    versionSpec = transformBasicVersionIntoRange(versionSpec);
```
kchodorow: You shouldn't need this.  VersionRange.createFromVersionSpec("1.2.3").containsVersion("1.2.4") is true.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import static com.google.devtools.build.workspace.maven.Rule.MAVEN_CENTRAL_URL;
+
+public class VersionResolver {
+
+  private final VersionHTMLScraper downloader;
```
kchodorow: Update to use aether.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
kchodorow: You shouldn't need to parse this at all, createFromSpec should handle any valid version.

petroseskinder:
per our offline discussion, i'm going to close this pull request.

PR #40, opened 2017-07-05
Implement transitive dependency resolution using Aether

petroseskinder:
In particular, given a set of artifacts, this CL produces a set of Bazel maven_jar rules. Transitive dependency resolution is achieved using aether. I rolled my own solution for serializing aether's dependency graph into a set of build rules. This is accompanied with a set of unit tests

This has not been integrated into the generate_workspace class. It's a slightly better use of time to first figure out how to parse maven artifacts from a set of pom files. 

The obscenely large size of the CL is due to changes to `generate_workspace.bzl`

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
@@ -1,0 +1,84 @@
+package com.google.devtools.build.workspace.maven;
+
+
+import com.google.common.collect.ImmutableList;
+import org.eclipse.aether.RepositorySystem;
+import org.eclipse.aether.RepositorySystemSession;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.eclipse.aether.collection.CollectRequest;
+import org.eclipse.aether.graph.Dependency;
+import org.eclipse.aether.graph.DependencyFilter;
+import org.eclipse.aether.graph.DependencyNode;
+import org.eclipse.aether.repository.RemoteRepository;
+import org.eclipse.aether.resolution.DependencyRequest;
+import org.eclipse.aether.resolution.DependencyResolutionException;
+import org.eclipse.aether.resolution.DependencyResult;
+import org.eclipse.aether.util.artifact.JavaScopes;
+import org.eclipse.aether.util.filter.DependencyFilterUtils;
+
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import java.util.function.Function;
+import java.util.logging.Logger;
+
+import static com.google.devtools.build.workspace.maven.AetherUtils.*;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Resolves transitive dependencies of maven artifacts using aether.
+ */
+public class ArtifactResolver {
+  private final RepositorySystem system;
+  private final RepositorySystemSession systemSession;
+  private final List<RemoteRepository> remoteRepositories;
+  private final List<Dependency> managedDependencies;
+
+  private final static Logger logger = Logger.getLogger(
+      MethodHandles.lookup().lookupClass().getName());
+
+  //TODO(petros): add support for managed dependencies, exclusions, and aliases.
+  public ArtifactResolver() {
+    this.system = newRepositorySystem();
+    this.systemSession = newRepositorySession(this.system);
+    this.remoteRepositories = ImmutableList.of(mavenCentralRepository());
+    this.managedDependencies = ImmutableList.of();
+  }
+
+  /**
+   * Given a set of coordinates, resolves the transitive dependencies, and then returns the root
+   * node to the resolved dependency graph. The root node is a sentinel node with direct edges
+   * on the artifacts users declared explicit on.
+   */
+  public DependencyNode resolveArtifacts(List<String> artifactCoords) {
+    DependencyRequest dependencyRequest = createDependencyRequest(artifactCoords);
+    DependencyResult dependencyResult;
+
+    try {
+      dependencyResult = system.resolveDependencies(systemSession, dependencyRequest);
+    } catch (DependencyResolutionException e) {
+      //FIXME(petros): This is very fragile. If even one artifact does not resolve, then no artifacts resolve.
+      logger.warning("Unable to resolve transitive dependencies: " + e.getMessage());
+      return null;
+    }
+    // root is a sentinel node whose direct children are the requested artifacts.
+    return dependencyResult.getRoot();
+  }
+
+  /**
+   * Given a list of artifact coordinates, creates an Aether DependencyRequest,
+   * i.e. a request all transitive dependencies. This request in particular,
+   * ignores any dependencies which are not scoped as Compile
+   */
+  private DependencyRequest createDependencyRequest(List<String> artifactCoords) {
+
+    Function<String, Dependency> coordinateToDependencyNode =
+        a -> new Dependency(new DefaultArtifact(a), JavaScopes.COMPILE);
+
+    List<Dependency> roots = artifactCoords.stream().map(coordinateToDependencyNode).collect(toList());
+
+    CollectRequest collectRequest = new CollectRequest(roots, managedDependencies, remoteRepositories);
+
+    DependencyFilter compileFilter = DependencyFilterUtils.classpathFilter(JavaScopes.COMPILE);
+    return new DependencyRequest(collectRequest, compileFilter);
+  }
+}
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
@@ -1,0 +1,82 @@
+package com.google.devtools.build.workspace.maven;
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.google.common.graph.GraphBuilder;
+import com.google.common.graph.MutableGraph;
+import org.eclipse.aether.graph.DependencyNode;
+import org.eclipse.aether.graph.DependencyVisitor;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Serializes a resolved dependency graph into a set of Bazel Build Rules.
+ */
+public class GraphSerializer {
+
+  public static Set<MavenJarRule> generateBuildRules(DependencyNode root) {
+    MutableGraph<MavenJarRule> graph = GraphBuilder.directed().allowsSelfLoops(false).build();
+    DependencyVisitor visitor = new AetherGraphTraverser(graph);
+    root.getChildren().forEach(c -> c.accept(visitor));
+    return graph.nodes();
+  }
+}
+
+/**
+ * Performs a depth first traversal of aether's dependency node graph, and
+ * constructs a guava graph representation of the aether graph, with each
+ * node being a Rule object.
+ */
+class AetherGraphTraverser implements DependencyVisitor {
+  private final Set<DependencyNode> visited;
+  private final MutableGraph<MavenJarRule> ruleGraph;
+  private final Map<DependencyNode, MavenJarRule> nodeRuleMap;
+
+  AetherGraphTraverser(MutableGraph<MavenJarRule> ruleGraph) {
+    this.visited = Sets.newHashSet();
+    this.ruleGraph = ruleGraph;
+    this.nodeRuleMap = Maps.newHashMap();
+  }
+
+  @Override
+  public boolean visitEnter(DependencyNode dependencyNode) {
+    if (dependencyNode == null || dependencyNode.getDependency().isOptional()) {
+      return false;
+    }
+    boolean isFirstVisit = visited.add(dependencyNode);
+    if (!isFirstVisit) {
+      return false;
+    }
+    MavenJarRule rule = getRule(dependencyNode);
+    for (DependencyNode child : dependencyNode.getChildren()) {
+      MavenJarRule childRule = getRule(child);
+      ruleGraph.putEdge(rule, childRule);
+    }
+    return true;
+  }
+
+  @Override
+  /* Adds parent and child pointers to nodes for later use by serializer. */
+  public boolean visitLeave(DependencyNode dependencyNode) {
+    MavenJarRule rule = getRule(dependencyNode);
+    ruleGraph.successors(rule).forEach(r -> r.addParent(rule));
+    ruleGraph.successors(rule).forEach(rule::addDependency);
+    return true;
+  }
+
+  /**
+   * Given a dependency node, either fetches a cached Rule object or
+   * constructs a new one.
+   */
+  private MavenJarRule getRule(DependencyNode node) {
+    MavenJarRule rule = nodeRuleMap.get(node);
+    if (rule != null) {
+      return rule;
+    }
+    rule = new MavenJarRule(node);
+    nodeRuleMap.put(node, rule);
+    ruleGraph.addNode(rule);
+    return rule;
+  }
+}
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
@@ -29,7 +29,7 @@ public class GraphSerializer {
  * node being a Rule object.
  */
 class AetherGraphTraverser implements DependencyVisitor {
-  private final Set<DependencyNode> visited;
+  private final Set<MavenJarRule> visited;
   private final MutableGraph<MavenJarRule> ruleGraph;
   private final Map<DependencyNode, MavenJarRule> nodeRuleMap;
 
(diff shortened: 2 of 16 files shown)


bazel-io:

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+public class ArtifactResolver {
+  private final RepositorySystem system;
+  private final RepositorySystemSession systemSession;
+  private final List<RemoteRepository> remoteRepositories;
+  private final List<Dependency> managedDependencies;
+
+  private final static Logger logger = Logger.getLogger(
+      MethodHandles.lookup().lookupClass().getName());
+
+  //TODO(petros): add support for managed dependencies, exclusions, and aliases.
```
kchodorow, line 39: Will this be handled "automatically" by Aether's machinery?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+   * on the artifacts users declared explicit on.
+   */
+  public DependencyNode resolveArtifacts(List<String> artifacts) {
+    DependencyRequest dependencyRequest = createDependencyRequest(artifacts);
+    DependencyResult dependencyResult;
+
+    try {
+      dependencyResult = system.resolveDependencies(systemSession, dependencyRequest);
+    } catch (DependencyResolutionException e) {
+      //FIXME(petros): This is very fragile. If even one artifact does not resolve, then no artifacts resolve.
```
kchodorow, line 59: You could remove the "return null".

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+      logger.warning("Unable to resolve transitive dependencies: " + e.getMessage());
+      return null;
+    }
+    return dependencyResult.getRoot();
```
kchodorow: What is the root when multiple dependencies are requested?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Given a list of artifact coordinates, creates an Aether DependencyRequest,
+   * i.e. a request all transitive dependencies. This request in particular,
+   * ignores any dependencies which are not scoped as Compile
+   */
+  private DependencyRequest createDependencyRequest(List<String> artifacts) {
```
kchodorow: Please rename artifacts -> "coordinates"

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/MavenJarRule.java
```diff
@@ -0,0 +1,145 @@
...
+  public boolean hasCustomRepository() {
+    List<RemoteRepository> repositories = node.getRepositories();
+    if (repositories == null || repositories.isEmpty() || repositories.size() > 1) {
+      return false;
+    }
+    return repositories.get(0).equals(mavenCentralRepository());
+  }
+
+  public String getRepository() {
+    System.out.println("REPOSITORY " + node.getRepositories());
```
kchodorow: Debugging

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Rule.java
```diff
@@ -16,18 +16,18 @@
...
+import javax.annotation.Nullable;
+import java.lang.invoke.MethodHandles;
 import java.util.Objects;
 import java.util.Set;
+import java.util.logging.Logger;
 
 /**
  * A struct representing the fields of maven_jar to be written to the WORKSPACE file.
  */
+//TODO(petros): Kill this after refactoring resolvers.
```
kchodorow, line 30: Why did you create a new class?

kchodorow (commented):
Generally looks good!

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
petroseskinder: The root is an empty sentinel node. Its children are the requested dependencies. This way dependencies are not resolved independently.

The key line is 

`CollectRequest collectRequest = new CollectRequest(roots, managedDependencies, remoteRepositories)`

Added a comment for clarity.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
petroseskinder: Little change. I return `dependencyResult.getRoot()` and `dependencyResult` is assigned within the try block. I would need to at some point, return null.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
petroseskinder: For the most part. Though, there is some massaging one needs to do with user input.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Rule.java
petroseskinder: It does effectively the same thing, except using aether classes. The original class uses maven versions of classes like Artifact and what not. I found that augmenting `Rule` class required a considerable amount of refactoring across the codebase. 

Since I am going to rip the resolvers out, I thought it would be a better use of time to keep both classes.

petroseskinder: This is the direction I am moving with [the next pull request](https://github.com/petroseskinder/migration-tooling/commit/14880cc251a907af40c8a696f4ac15ed7a4368d3#diff-376ae4a942b0ed1920d7f8b28cf11751), as far as ripping out the Resolver

petroseskinder:
I'm not sure why github is stating there are conflicts within generate_workspace.bzl
As of commit 9880260, there are no conflicts in generate_workspace.bzl.  😞

kchodorow:
Checking out this branch, I get:

```
$ git fetch git://github.com/petroseskinder/migration-tooling.git version-fix
remote: Counting objects: 75, done.
remote: Total 75 (delta 16), reused 16 (delta 16), pack-reused 58
Unpacking objects: 100% (75/75), done.
From git://github.com/petroseskinder/migration-tooling
 * branch            version-fix -> FETCH_HEAD
03:17:13 PM ~/gitroot/migration-tooling (master)$ git reset --hard FETCH_HEAD
HEAD is now at bc0918e Add statement in README acknowledging aether
03:17:22 PM ~/gitroot/migration-tooling (master)$ bazel run //generate_workspace -- -a org.apache.maven:maven-artifact:3.5.0
ERROR: /home/kchodorow/gitroot/migration-tooling/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/BUILD:1:1: no such package '@org_eclipse_aether_aether_util//jar': The repository could not be resolved and referenced by '//generate_workspace/src/main/java/com/google/devtools/build/workspace/maven:maven'.
ERROR: Analysis of target '//generate_workspace:generate_workspace' failed; build aborted.
INFO: Elapsed time: 1.802s
ERROR: Build failed. Not running target.
```

petroseskinder:
Ehh. I see. My apologies. This is from merging the generate_workspace.bzl. It should be fine now. I squashed a number of commits and regenerated the file. 

After the repository rule is merged, it makes sense to add this to the gitignore.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
kchodorow: It looks like you can get the dependency result from the exception, so you can still return it (and show which dependencies did not resolve).

Please test.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
```diff
@@ -0,0 +1,82 @@
...
+    this.nodeRuleMap = Maps.newHashMap();
+  }
+
+  @Override
+  public boolean visitEnter(DependencyNode dependencyNode) {
+    if (dependencyNode == null || dependencyNode.getDependency().isOptional()) {
+      return false;
+    }
+    boolean isFirstVisit = visited.add(dependencyNode);
+    if (!isFirstVisit) {
```
kchodorow: In my experience, Maven/Aether does not define equals() methods, so probably `new DependencyNode() != new DependencyNode()`.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
```diff
@@ -0,0 +1,82 @@
...
+      ruleGraph.putEdge(rule, childRule);
+    }
+    return true;
+  }
+
+  @Override
+  /* Adds parent and child pointers to nodes for later use by serializer. */
+  public boolean visitLeave(DependencyNode dependencyNode) {
+    MavenJarRule rule = getRule(dependencyNode);
+    ruleGraph.successors(rule).forEach(r -> r.addParent(rule));
```
kchodorow, line 63: I don't think we want the parents.  I was parsing them for pom.xml files because I needed the variables they defined, I think.  But we don't actually want their dependencies.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/GraphSerializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    Set<MavenJarRule> rules = GraphSerializer.generateBuildRules(sentinel);
+
+    MavenJarRule ruleA = new MavenJarRule(nodeA);
+    MavenJarRule ruleA1 = new MavenJarRule(nodeA1);
+    MavenJarRule ruleB = new MavenJarRule(nodeB);
+    MavenJarRule ruleC = new MavenJarRule(nodeC);
+    addDependency(ruleA, ruleA1);
+
+    assertRuleSetContainsExactly(rules, ruleA, ruleA1, ruleB, ruleC);
```
kchodorow, line 67: Add tests for duplicate children:

```
DependencyNode node1 = dependencyNode("a:a:1");
DependencyNode node2 = dependencyNode("a:a:1");
```

petroseskinder: Let me make sure I understand your request. You are asking to test that we never get a rule like
```
A --> {"a:a:1", "a:a:1"}
```
If so, I have added a test within the AetherGraphTraversalTest. That is where this condition is enforced.

petroseskinder: Let me make sure I understand your request. You are asking to test that we never get a rule like
```
A --> {"a:a:1", "a:a:1"}
```
If so, I have added a test within the AetherGraphTraversalTest. That is where this condition is enforced.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
petroseskinder: Disappointing they never override the equals() method. 

This is a potential performance issue, since we don't want to traverse/accept a node multiple times. It doesn't affect correctness.

I suspect that object.equals() may suffice for identifying equal nodes. 

This suspicion derives from Aether's [TreeDependencyVisitor](https://github.com/sonatype/sonatype-aether/blob/master/aether-util/src/main/java/org/sonatype/aether/util/graph/TreeDependencyVisitor.java) class. 

```java
public boolean visitEnter( DependencyNode node )
    {
        boolean visited = visitedNodes.put( node, Boolean.TRUE ) != null;

        visits.push( Boolean.valueOf( visited ) );

        if ( visited )
        {
            return false;
        }

        return visitor.visitEnter( node );
    }

```
[Related tests](https://github.com/sonatype/sonatype-aether/blob/master/aether-util/src/test/java/org/sonatype/aether/util/graph/TreeDependencyVisitorTest.java)

I will include a TODO, and investigate this further.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
petroseskinder: Don't we want to include in the `generate_workspace.bzl` who requested a given artifact?
```python
# something:something.id requested this
native.maven_jar(
)
```
It is much easier doing this bookkeeping here, rather than having to compute this information later.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/ArtifactResolver.java
petroseskinder: Noted. Can we separate this out to another pull request? 
This pull request doesn't alter the functionality of generate_workspace. I would like to delay until I remove the existing Resolver.

petroseskinder:
> It looks like you can get the dependency result from the exception, so you can still return it (and show which dependencies did not resolve).
>
> Please test.

Noted. Can we separate this out to another pull request? Specifically, I would prefer this be in the same PR that rips out Resolver.java

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
kchodorow: As I suspected, this fails:

```
    DependencyNode n1 = dependencyNode("x:y:1");
    DependencyNode n2 = dependencyNode("x:y:1");
    assertThat(n1).isEqualTo(n2);
```

Can you just make `visited` a set of MavenJarRules? That avoids this whole issue.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
kchodorow: Oh right, forgot about that.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/GraphSerializerTest.java
kchodorow: No, I mean if we get `x:x:1 -> {y:y:1 -> {a:a:1}, z:z:1 -> {a:a:1}}` we don't end up with two `a:a:1` nodes.

kchodorow (commented):
Okay.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/GraphSerializerTest.java
petroseskinder: I see. Test for that can be found [here](https://github.com/bazelbuild/migration-tooling/pull/40/commits/69d01192083dfbb8587c127b8753527e441420ff#diff-681fe7c9eb94648ae236eee7b1e13a5dR70)

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/GraphSerializer.java
petroseskinder: Changed.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/GraphSerializerTest.java
kchodorow: What I was looking for was:

```
    DependencyNode sentinel = dependencyNode("dummy:dummy:0");
    DependencyNode nodeA = dependencyNode("a:a:1");
    DependencyNode nodeB = dependencyNode("b:b:1");
    DependencyNode child1 = dependencyNode("child:child:1");
    DependencyNode child2 = dependencyNode("child:child:1");

    sentinel.setChildren(ImmutableList.of(nodeA, nodeB));
    nodeA.setChildren(ImmutableList.of(child1));
    nodeB.setChildren(ImmutableList.of(child2));

    Set<MavenJarRule> rules = GraphSerializer.generateBuildRules(sentinel);

    MavenJarRule ruleA = new MavenJarRule(nodeA);
    MavenJarRule ruleB = new MavenJarRule(nodeB);
    MavenJarRule childRule = new MavenJarRule(child1);
    addDependency(ruleA, childRule);
    addDependency(ruleB, childRule);

    assertRuleSetContainsExactly(rules, ruleA, ruleB, childRule);
```

Which fails.

petroseskinder: Yes, failure is expected. We should not confront a situation like this (otherwise we have much more worrisome issues). 

Currently, to construct this graph, we are using aether's built in dependency resolution mechanism. I think it is safe to assume that aether's graph will not have situations like this. As you have pointed out, DefaultDependencyNode, the node used by aether, does not override its equals method. So the only way to assert equality is by object's `equals()` method. 

If the above situation were to occur, things would be broken/faulty much earlier in the pipeline. Were we to define our own dependency resolution mechanism, we could implement a DependencyNode that actually overrides `equals()`. 

Thoughts?

kchodorow: Ok.

kchodorow (approved):

kchodorow:
Can you resolve the generate_workspace.bzl conflicts & I'll merge?

kchodorow:
Jenkins, test this please.

petroseskinder:
jenkins says yay! :)

PR #42, opened 2017-07-07
Automatically create output directory if it doesn't exist #39

petroseskinder:
This is in reference to #39. To prevent future regressions, I added a relevant unit test. 
The large commit size is 99% due to generate_workspace.bzl

Reviewers: @kchodorow , @katre

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
@@ -15,11 +15,14 @@
 package com.google.devtools.build.workspace.output;
 
 import com.google.devtools.build.workspace.maven.Rule;
+
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.invoke.MethodHandles;
+import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.logging.Logger;
 
@@ -32,27 +35,33 @@ public class BzlWriter extends AbstractWriter {
       MethodHandles.lookup().lookupClass().getName());
 
   private final String[] argv;
-  private final Path generatedFile;
+  private final File generatedFile;
 
-  public BzlWriter(String[] argv, String outputDirStr) {
+  public BzlWriter(String[] argv, Path outputDir) {
     this.argv = argv;
-    this.generatedFile = Paths.get(outputDirStr).resolve("generate_workspace.bzl");
+    this.generatedFile = new File(outputDir.toString(), "generate_workspace.bzl");
   }
 
   @Override
   public void write(Collection<Rule> rules) {
-    try (PrintStream outputStream = new PrintStream(generatedFile.toFile())) {
-      writeBzl(outputStream, rules);
+    try {
+      createParentDirectory(generatedFile.getAbsoluteFile().toPath());
     } catch (IOException e) {
+      logger.severe("Could not create directories for " + generatedFile + ": " + e.getMessage());
+      return;
+    }
+    try (PrintStream outputStream = new PrintStream(generatedFile.getAbsoluteFile())) {
+      writeBzl(outputStream, rules);
+    } catch (FileNotFoundException e) {
       logger.severe("Could not write " + generatedFile + ": " + e.getMessage());
       return;
     }
-    System.out.println("Wrote " + generatedFile.toAbsolutePath());
   }
 
   private void writeBzl(PrintStream outputStream, Collection<Rule> rules) {
     writeHeader(outputStream, argv);
     outputStream.println("def generated_maven_jars():");
+
     if (rules.isEmpty()) {
       outputStream.println("  pass\n");
     }
@@ -70,4 +79,14 @@ public class BzlWriter extends AbstractWriter {
       outputStream.println(formatJavaLibrary(rule, "native.java_library", "  "));
     }
   }
+
+  /** Creates parent directories if they don't exist */
+  private static void createParentDirectory(Path generatedFile) throws IOException {
+    Path parentDirectory = generatedFile.toAbsolutePath().getParent();
+    if (Files.exists(parentDirectory)) {
+      return;
+    }
+    Files.createDirectories(parentDirectory);
+  }
+
 }
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
@@ -14,15 +14,16 @@
 
 package com.google.devtools.build.workspace.output;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.devtools.build.workspace.maven.Rule;
 
-import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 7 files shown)


bazel-io:

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
```diff
@@ -70,4 +75,14 @@ private void writeBzl(PrintStream outputStream, Collection<Rule> rules) {
...
     }
   }
+
+  /** Creates parent directories if they don't exist */
+  private static void createParentDirectory(Path generatedFile) throws IOException {
+    Path parentDirectory = generatedFile.toAbsolutePath().getParent();
+    if (Files.exists(parentDirectory)) {
+      return;
+    }
+    File file = new File(parentDirectory.toString());
```
kchodorow: Remove

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
```diff
@@ -41,7 +43,9 @@ public BzlWriter(String[] argv, String outputDirStr) {
 
   @Override
   public void write(Collection<Rule> rules) {
-    try (PrintStream outputStream = new PrintStream(generatedFile.toFile())) {
+    try {
```
kchodorow: Move the createParentDirectory out of the try/catch and continue to use try-with-resources.

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
```diff
@@ -35,18 +42,33 @@
...
+  /** Ensures that it automatically creates output directory if they do not exist */
+  @Test
+  public void automaticallyCreateParentDirectories() throws Exception {
+    FileSystem fs = Jimfs.newFileSystem(Configuration.unix());
+    Path root = fs.getPath("root");
+    createDirectories(root);
+
+    Path outputDir = root.resolve("child/directory/");
+
+    BzlWriter writer = new BzlWriter(new String[]{}, outputDir.toString());
```
kchodorow: Does this actually work? I'm surprised you don't need to pass in the Path.

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
```diff
@@ -35,18 +42,33 @@
...
+    writer.write(createRules("x:y:1.2.3"));
+
+    assertThatFileExists(outputDir.resolve("generate_workspace.bzl"));
+  }
```
kchodorow: Add a test for if the directories cannot be created.

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
petroseskinder: You pass the path of the directory that you want your bzl file in. 
That is `outputDir.toString()` 

If you look at the issue, they only pass the output directory.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
petroseskinder: It's going to throw an IOException as well. You would prefer two separate try-catch blocks to one?

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
kchodorow: But you're using an in-memory filesystem in the test and BzlWriter will try to write to /root/child/directory (IIUC).

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
kchodorow: Yeah, because it's different errors (creating directories vs. streaming content) and this way you don't have to worry about closing the output stream on failure.

petroseskinder: This makes sense. Changes made.

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
petroseskinder: I see what you mean. 

I passed in the path directly to the BzlWriter and use a java.nio.File. Current changes definitely work. I tested by asserting that a file that should exists and that a file that shouldn't exist does not exist. 

Also, I've been using `root/child/directory` rather than `/root/child/directory` Does the name root suggest the latter?

generate_workspace/src/test/java/com/google/devtools/build/workspace/output/BzlWriterTest.java
petroseskinder: Done. The additional test tries outputting to `/root/something/` which fails since Jimfs doesn't allow creating files using absolute paths. 

What exactly are you wanting to test? I simply assert that `/root/something/generate_workspace.bzl` does not exist. However, the nonexistence of this file could be for a host of reasons. For example, if we change BzlWriter to write to `fancy_generate_workspace.bzl` this test would still pass, but it wouldn't communicate anything of value.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
```diff
@@ -32,27 +35,33 @@
...
+      logger.severe("Could not create directories for " + generatedFile + ": " + e.getMessage());
+      return;
+    }
+    try (PrintStream outputStream = new PrintStream(generatedFile.getAbsoluteFile())) {
+      writeBzl(outputStream, rules);
+    } catch (FileNotFoundException e) {
       logger.severe("Could not write " + generatedFile + ": " + e.getMessage());
       return;
     }
-    System.out.println("Wrote " + generatedFile.toAbsolutePath());
```
kchodorow: Why is this removed?

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
```diff
@@ -32,27 +35,33 @@
       MethodHandles.lookup().lookupClass().getName());
 
   private final String[] argv;
-  private final Path generatedFile;
+  private final File generatedFile;
```
kchodorow: Why change this?

kchodorow (commented):
Please push new commits instead of squashing when you address comments. That way I can see the progression (and GitHub won't let me respond to "outdated" comments inline).

> What exactly are you wanting to test? I simply assert that /root/something/generate_workspace.bzl does not exist. However, the nonexistence of this file could be for a host of reasons. For example, if we change BzlWriter to write to fancy_generate_workspace.bzl this test would still pass, but it wouldn't communicate anything of value.

I just wanted to test that if createDirectories throws an IOException, we recover gracefully.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
petroseskinder: That's an error on my part. I will amend that.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
petroseskinder: Solely for unit tests with jimfs. 

The behavior of `createParentDirectory` differed in the in-memory filesystem between when I used `new File()` and when I used `somePath.toFile()` With the `new File()`everything (tests and actual tool) performed as desired. Using the Path, the tests would fail to create the directories. 

The only real distinction is that with one, I am performing:

```
createParentDirectory(outputFile);
```
and with the other I am doing
```
File file = new File(ouputDir.toString(), "generate_workspace.bzl");
...
createParentDirectory(file.getAbsoluteFile().toPath());
```

petroseskinder (commented):
> Please push new commits instead of squashing when you address comments. That way I can see the progression (and GitHub won't let me respond to "outdated" comments inline).

Duly noted.

kchodorow:
Jenkins, test this please.

generate_workspace/src/main/java/com/google/devtools/build/workspace/output/BzlWriter.java
kchodorow: That worries me.  Do you know why somePath.toFile() was failing?

petroseskinder: It was unable to create the file at `root/child/directory/generate_workspace.bzl`.
It would attempt to create the file at 
```
bazel-out/local-fastbuild/bin/.../BazelWriterTest.runfiles/__main__/
```
Then it would find that `root/child/directory/generate_workspace.bzl` was not created.

The IOException occurred when trying to create the PrintStream.

kchodorow: Hmm. There's not much point of using jimfs if it won't let us use an in-memory filesystem anyway.  Please modify this to either be all in memory or all on disk (for tests).

petroseskinder: Here is a less hand-wavy answer for why we need to use the File.

With the File approach, we create the PrintStream, using `file.getAbsoluteFile()`. It ends up writing to `~/execroot/migration-tooling/_tmp/BzlWriterTest_1/generate_workspace.bzl` 

If I wanted to replicate the behavior of `file.getAbsoluteFile()`we could use `path.toAbsolutePath().toFile()` rather than `path.toFile()` This would ensure we actually write to the jimfs file. 

However, jimfs.JimfsPath throws an UnsupportedOperationException.

kchodorow: I maintain that there's not much point in using jimfs then, if we have to drop to the normal filesystem anyway.

petroseskinder: > Hmm. There's not much point of using jimfs if it won't let us use an in-memory filesystem anyway. Please modify this to either be all in memory or all on disk (for tests).

Wait, check the follow up comment. using `path.toFile()` didn't use the in-memory filesystem. We needed the absolute path. From what I understand, using `file.getAbsoluteFile()` let us use the in-memory file system. Each of the files were written to the bazel-sandbox's _tmp subdirectory, and were deleted after test completion.

petroseskinder: > I maintain that there's not much point in using jimfs then, if we have to drop to the normal filesystem anyway.

I see. Let me see what I can do then.

kchodorow: That's not an in-memory filesystem, it's just an on-disk filesystem that gets cleaned up.

petroseskinder:
Jenkins, test this please.

kchodorow:
LGTM.

PR #44, opened 2017-07-18
Don't include dependency management specifications as deps

### LINKED ISSUE #41: --artifact is pulling in unrelated artifacts
I ran this on a fresh clone of `migration-tooling`:
```
bazel run //generate_workspace -- --output_dir /tmp --artifact=com.google.guava:guava:22.0
```
and it wrote a `generate_workspace.bzl` with 138 dependencies:

```
% grep -A 1 native.maven_jar /tmp/generate_workspace.bzl | grep name | wc -l       
138

% grep -A 1 native.maven_jar /tmp/generate_workspace.bzl | grep name | sort | head
      name = "ant_ant",
      name = "asm_asm",
      name = "backport_util_concurrent_backport_util_concurrent",
      name = "cglib_cglib",
      name = "classworlds_classworlds",
      name = "com_google_code_findbugs_jsr305",
      name = "com_google_errorprone_error_prone_annotations",
      name = "com_google_guava_guava",
      name = "com_google_j2objc_j2objc_annotations",
      name = "com_jcraft_jsch",
```

When I run the similar command in the main bazel repo, I only get 5:

```
% cd /tmp/bazel

% bazel run //src/tools/generate_workspace -- --output_dir /tmp --artifact=com.google.guava:guava:22.0

% grep -A 1 maven_jar /tmp/WORKSPACE | grep name | wc -l                                                                                                                       
5
```

```
% bazel version
Build label: 0.5.2
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Jun 27 13:27:03 2017 (1498570023)
Build timestamp: 1498570023
Build timestamp as int: 1498570023
```

kchodorow:
My previous implementation found the transitive dependencies of any dependency mentioned in the DependencyManagement section and added them all as dependencies.  This was very incorrect and very bloated.

Now that the generate_workspace.bzl has been de-bloated, I've switched the BUILD files to using the generated java_libraries and ran unused_deps on everything.

Fixes #41.


kchodorow:
AFAICT everything passed ("Executed 7 out of 7 tests: 7 tests pass.") so I'm not sure why the pipeline says it fails... summoning @damienmg.

petroseskinder:
[Jenkins says](http://ci.bazel.io/blue/organizations/jenkins/PR%2Fmigration-tooling/detail/migration-tooling/16/pipeline/248):
```
groovy.lang.MissingMethodException: No signature of method: build.bazel.ci.BazelUtils.copyCommands() is applicable for argument types: (java.util.ArrayList, java.util.HashMap) values: [[], [name:test.xml, uri:file:///home/ci/.cache/bazel/_bazel_ci/a123b34acc5de28efdcbead8be7ebf66/execroot/__main__/bazel-out/local-fastbuild/testlogs/generate_workspace/src/test/java/com/google/devtools/build/workspace/output/WorkspaceWriterTest/test.xml]]
```
Did the WorkspaceWriterTests also pass?

kchodorow:
It looks it (`//generate_workspace/src/test/java/com/google/devtools/build/workspace/output:WorkspaceWriterTest PASSED in 1.0s`), so I'm guess a pipeline misconfiguration.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/BUILD
```diff
@@ -3,20 +3,17 @@ java_library(
     srcs = glob(["*.java"]),
     visibility = ["//generate_workspace/src:__subpackages__"],
     runtime_deps = [
-        "@org_codehaus_plexus_plexus_interpolation//jar",
-        "@org_codehaus_plexus_plexus_utils//jar",
+        "//third_party:org_codehaus_plexus_plexus_interpolation",
```
petroseskinder, line 6: I see that you have changed all of the external dependencies to use third party java_libraries, rather than the maven_jar rules. 

Does this change mean that we can remove the following lines from the workspace file
```
load('//:generate_workspace.bzl', 'generated_maven_jars')
generated_maven_jars()
```
Also, out of curiosity, why have you chosen to include the java_library targets rather than the maven third party ones?

damienmg:
Sorry I pushed a wrong config, just retrigger the build and it should be
alright

kchodorow:
Cool, np. Test this please!

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
```diff
@@ -328,11 +333,37 @@ private boolean addArtifact(Rule dependency, String parent) {
...
+      logger.warning(
+          "Error parsing version " + versionRestriction + ": " + e.getLocalizedMessage());
+      // So that this isn't logged over and over.
+      restriction.remove(artifactName);
+      return;
+    }
+    if (!versionRange.containsVersion(new DefaultArtifactVersion(dependency.version()))) {
+      try {
+        dependency.setVersion(
+            resolveVersion(dependency.groupId(), dependency.artifactId(), versionRestriction));
```
petroseskinder, line 363: Can you include a TODO or FIXME statement here? 

This resolveVersion is the incorrect one defined [here](https://github.com/bazelbuild/migration-tooling/pull/44/commits/17b568cbd53e4c67b754ee24c331fed51a81bc7c#diff-cc04e6e62b5271603d5f3061cf8fc024L102). Using that function ensures we are consistent about how we resolve versions/version ranges, but it is still incorrect. I think its important we make note of this as a reminder.

If you recall, we currently interpret "[1.0, )" as 1.0 rather than the highest available, and fail for things like "(3, 5)" which we shouldn't. This means we will erroneously receive InvalidArtifactCoordinateExceptions.

damienmg:
BTW this looks like the PR I was waiting for, is it?

kchodorow:
I have no idea, what were you waiting for?

damienmg:
I had that dependencies explosion trying to pull test deps for testing
Jenkins pipeline (the kind of test that would have prevented the error
above :p)

kchodorow:
Oh yeah! Well, here's the fix.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/ResolverTest.java
```diff
@@ -145,6 +145,26 @@ public void dependencyManagementWins() throws Exception {
   }
 
   @Test
```
petroseskinder, line 152: Can you include a test when the depmgt is a version range. For example:

```java
// basic example
Dependency managed = getDependency("a:b:[1.0, 4.0)");
Dependency version = getDependency("a:b:2.0");

// gnarlier example
Dependency managed = getDependency("a:b:[1.0, 4.0)");
Dependency version = getDependency("a:b:[1.2, 3.0]");
```
Currently, I am unsure how that will behave. I think this is a useful way to document how we resolve versions.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/BUILD
kchodorow: No, we can't remove those lines from the WORKSPACE file.  They are what actually import the Maven jars!  The //third_party targets just depend on the maven_jars.  

I changed it to use the libraries because they are user-friendlier: suppose there is a Maven artifact `a:a:1.0` that depends on `b:b:1.0`.  If you're depending on `@a_a//jar`, your `java_library` has to also depend on `@b_b//jar` or you could get an error when your code calls something in `@a_a//jar` and `@b_b//jar` isn't in the classpath.  

The `java_library` definitions are generated from `a:a:1.0`'s list of dependencies, so depending on `//third_party:a_a` gives you the dependency on `@b_b//jar`, too.

petroseskinder: I see. Thank you for the clarification, particularly the example.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
kchodorow: Done.

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/ResolverTest.java
kchodorow: Done.

petroseskinder:
LGTM. Tried it out with `com.google.guava:guava:22.0` and it checks out. No unnecessary dependencies.

petroseskinder:
@buchgr any status on the review

kchodorow:
Given I've gotten an LGTM from @petroseskinder and it's been a couple days, going to submit.

PR #46, opened 2017-07-31
Implements transitive maven jar repository rule.

petroseskinder:
This pull request provides an initial implementation of the `transitive_maven_jar` rule. The repository rule self hosted and generated the generate_workspace.bzl found in this CL.

Design considerations and additional slated functionality is discussed in [this design document](https://docs.google.com/document/d/1U3qDaya88BEBmSoA4z4uqFPP7aNLA1jJJ2PmhFT5lNQ/edit?usp=sharing).

Provides a working solution for the upgrade problem discussed in #20 

@kchodorow


bazel-io:

README.md
```diff
@@ -32,5 +32,38 @@ generated_java_libraries()
...
+After the initial migration, one can use the `transitive_maven_jar` repository rule
+to manage one's external maven dependencies.
+
+First, add the following to your `WORKSPACE` file: 
+
+```python
+http_archive(
+	name = "trans_maven_jars",
+	url = "https://github.com/petroseskinder/migration_tooling/releases/download/test/generate_workspace_deploy.jar",
+	type = "jar",
```
kchodorow: You don't actually need the type line.

README.md
```diff
@@ -32,5 +32,38 @@ generated_java_libraries()
 Then you can access any of the Java library targets in `generate_workspace.bzl`,
 which export each jar's dependencies as well as the jar itself.
 
+## `transitive_maven_jar` 
+
+After the initial migration, one can use the `transitive_maven_jar` repository rule
```
kchodorow: I'd describe this more as an alternative.  Maybe something like:

> The `transitive_maven_jar` repository rule exposes the specified Maven jars, as well as their transitive dependencies, under one target.  This avoids running `generate_workspace` when dependencies change, at the cost of re-fetching them all and having a less fine-grained dependency structure.

transitive_maven_jar/BUILD
```diff
@@ -0,0 +1,4 @@
+package(
```
kchodorow: Remove the content, leave an empty file.  .bzl files don't need to be marked public.

transitive_maven_jar/transitive_maven_jar.bzl
```diff
@@ -0,0 +1,55 @@
...
+	_validate_coordinates(rctx)
+	arguments = _create_arguments(rctx)
+
+	# obtain the generate workspace binary
+	jar_path = rctx.path('generate_workspace_deploy.jar')
+	rctx.download(BINARY_URL, output=jar_path)
+
+	# execute the command
+	result = _execute(rctx, "java -jar %s %s" % (jar_path, arguments))
+	print(result.stdout) # TODO(petros): print to console directly
```
kchodorow: You can specify `quiet = False` in rctx.execute to print to console as it goes.

transitive_maven_jar/transitive_maven_jar.bzl
```diff
@@ -0,0 +1,55 @@
...
+	return rctx.execute(["bash", "-c", command_string], timeout=rctx.attr.timeout)
+
+def _transitive_maven_jar_impl(rctx):
+	_check_dependencies(rctx)
+	_validate_coordinates(rctx)
+	arguments = _create_arguments(rctx)
+
+	# obtain the generate workspace binary
+	jar_path = rctx.path('generate_workspace_deploy.jar')
+	rctx.download(BINARY_URL, output=jar_path)
```
kchodorow: You don't need to download this twice, use the one that was previously declared in the WORKSPACE file.

transitive_maven_jar/transitive_maven_jar.bzl
```diff
@@ -0,0 +1,55 @@
...
+def _validate_coordinates(rctx):
+	coordinates = rctx.attr.artifacts
+	for coord in coordinates:
+		if _validate_coordinate_length(coord) == False:
+			fail("Invalid coordinate %s. Generally formatted as \"group:artifact:version\"" % coord)
+	return True
+
+def _create_arguments(rctx):
+	#TODO(petros): pass output directory as an attribute.
+	arguments = ['-a ' + artifact for artifact in rctx.attr.artifacts]
```
kchodorow: Might as well go for readability here and do `--artifact`.

You should be able to use [`rctx.path()`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#path) to get the output dir.

README.md
petroseskinder: Sounds good. One question. How does it lead to a less fine-grained dependency structure? The repository rule is simply acting as a wrapper around `generate_workspace` so it should have no influence on the dependency structure.

README.md
petroseskinder: done. 

I believe this is a mistake. The url should point to the migration tool's zipped archive rather than the generate_workspace_deploy jar. In addition, do you plan on pushing the generate workspace tool to maven central.

transitive_maven_jar/BUILD
petroseskinder: done.

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: Yeah, I have been consistently getting the following error:

```
rctx.execute(["bash", "-c", command_string], ti.... , ....)
too many arguments, in method execute(list, int timeout, bool quiet) of 'repository_ctx`.
```
For context, I am passing the executing the following command
```
rctx.execute(["bash", "-c", command_string], timeout = rctx.attr.timeout, quiet = False)
```
Strange as this is not an issue when I execute
```
rctx.execute(["bash", "-c", command_string], environment = {}, timeout = rctx.attr.timeout)
```

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: > Might as well go for readability here and do --artifact.

done.
> You should be able to use [`rctx.path()`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#path) to get the output dir.

Oh yeah. I think the TODO was mislabeled. I was more interested in having the `generate_workspace.bzl` file be outputted in a configurable location.

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: wip. 

Unclear what I am downloading with the http_archive.

petroseskinder (commented):
Thank you. There were a few points of confusion. In particular, whether the http archive should point to the zipped migration tools or the generate_workspace binary.

README.md
kchodorow: Faulty thinking on my part: I was thinking that if you updated a dep, it would have to redownload everything.

However, thinking about this further, `transitive_maven_jar(name = "foo")` should be able to copy generate_workspace.bzl file in /ob/external/foo, so people should be able to do:

```
transitive_maven_jar(
    name = "foo",
    ...
)

load('@foo//:generate_workspace.bzl', 'generated_maven_jars')
generated_maven_jars()
```

Please add that to the code example.

petroseskinder: Done. Thank you.

README.md
```diff
@@ -32,5 +32,47 @@ generated_java_libraries()
...
+```
+The `transitive_maven_jar` rule will resolve the transitive dependencies for the
+specified artifacts and then it will generate a `generate_workspace.bzl` which
+you must load in your `WORKSPACE` file.
+
+After the files are generated, add the following lines into your WORKSPACE file to load the `maven_jar` rules
+
+```
+load("@dependencies//:generate_workspace.bzl", "generated_java_libraries")
+generated_java_libraries()
```
kchodorow: Remove, this doesn't go in the WORKSPACE file.

diff
```diff
@@ -0,0 +1,154 @@
+diff --git a/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/AbstractWriter.java b/generate_workspace/src/main/java/com/google/devtools/build/workspace/output/AbstractWriter.java
```
kchodorow: Remove?

settings_command
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
kchodorow: Remove

transitive_maven_jar/transitive_maven_jar.bzl
kchodorow: `quiet` is a fairly recent addition to Bazel, what version are you running?

transitive_maven_jar/transitive_maven_jar.bzl
```diff
@@ -0,0 +1,59 @@
...
+	result = _execute(rctx, "java -jar %s %s" % (jar_path, arguments))
+	print(result.stdout) # TODO(petros): print to console directly
+	rctx.file('%s/BUILD' % rctx.path('test'), '', False)
+
+transitive_maven_jar = repository_rule(
+		implementation = _transitive_maven_jar_impl,
+		attrs = {
+			"artifacts" : attr.string_list(default = [], mandatory = True),
+			"timeout" : attr.int(default = MAX_TIMEOUT),
+			"output_directory" : attr.string(default = "")
```
kchodorow: What is this for?

diff
petroseskinder: 😞 removing. Did `git add .`

settings_command
petroseskinder: done.

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: this is for where you want the `generate_workspace.bzl` file to be generated to.

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: 0.4.5-homebrew. I'll upgrade. Thank you.

transitive_maven_jar/transitive_maven_jar.bzl
petroseskinder: I can just remove this. The tool will indicate where the file is, and users can just copy or symlink it themselves manually.

README.md
petroseskinder: done

transitive_maven_jar/transitive_maven_jar.bzl
kchodorow: They don't have to. The repository rule itself creates the directory (`ob/external/<rule name>`) and the repo rule can (should) copy the generate_workspace.bzl file to `ob/external/<rule name>/generate_workspace.bzl`.

petroseskinder: > the repo rule can (should) copy the generate_workspace.bzl file to `ob/external/<rule name>/generate_workspace.bzl`

what is ob in this context?

As it stands, it already creates the file in `_bazel_$USER/hash/external/<rule name>/generate_workspace.bzl` I believe that is what you are referring to.

What I was (unclearly) referring to was whether it should create a symlink within the WORKSPACE.

kchodorow: ob == "output base" (same as the `/.../_bazel_$USER/hash` path). I don't think we need to create a symlink in the main workspace (which I think is what you're referring to), which is nicer for the user (they can just add stuff to their WORKSPACE file and start building).

petroseskinder: This makes sense. Take a look at the changes [here](https://github.com/bazelbuild/migration-tooling/pull/46/commits/2172ca910df138d90164ffcdf6c23886cbba666d). With this change, I believe users can just load the file with `load("@name_of_deps//:generate_workspace.bzl", "generated_maven_jars")`

The only issue that remains after this is with us constantly downloading the java binary.

petroseskinder:
@kchodorow I resolved the issue of constantly downloading the generate_workspace tool by adding the deploy jar in the `transitive_maven_jar` directory. 

It works, but by no means is it optimal. I would rather not check a binary into the repository. However, this approach allows us to point users to an http_archive of _just_ the `transitive_maven_jar` repository rule. That is, without leaking any information of how the tool resolves dependencies. The only annoyance is on our end with maintaining the jar. On that front, Alpha suggested creating a regular old rule to package that directory into a zip file.

kchodorow:
I'd rather have the generate_workspace deploy jar be downloaded. Other than that, this looks good to go.

petroseskinder:
> I'd rather have the generate_workspace deploy jar be downloaded.

Why is that? 

I don't feel strongly either way, but downloading the jar seems a little counter intuitive to me.

From a user's perspective, packaging them together is clearly preferable. 

From our standpoint, we have little to gain by downloading the deploy jar.  At least, how I have interpreted it. The two general approaches we could take for downloading would be to download from within the repository rule as I did before or to have the user use a `http_archive`. 

The first is obviously bad. 

The second (using `http_archive`) is unnecessarily clunky for users, as they have to manually write the `http_archive`. From our end, we would still need to host the executable jar _**somewhere**_. That is until Ulf or Damian decide it is appropriate to expose bazel under `bazel-tools`, we can't just use the `http_archive` of the migration-tooling repository (i.e. `github.com/bazelbuild/migration-tooling/archive/master.zip`, [non-working example here](https://github.com/petroseskinder/migration-tooling/commit/3003f6c4b76a0215561a5f5997ac9303de505199)). 

If we are going manually host and update the jar anyways, why not package it together with the repository rule, as I have done, and make users lives easier? In addition, the repository rule is strongly dependent on the version of `generate_workspace` it calls.  Why give users the flexibility to potentially break things?

kchodorow:
I was thinking it would be better to have it be separate, so it would be easy to swap out a different version. However, I'm willing to merge it in as-is and see how it goes.

But: I was thinking about having //generate_workspace depend on transitive_maven_jar which runs generate_workspace and that seems like a bad idea.  It'll be annoying to bootstrap that if any step breaks.  Thus, please revert the WORKSPACE to using plain ol' generate_workspace.bzl.

petroseskinder:
> But: I was thinking about having //generate_workspace depend on transitive_maven_jar which runs generate_workspace and that seems like a bad idea. It'll be annoying to bootstrap that if any step breaks. Thus, please revert the WORKSPACE to using plain ol' generate_workspace.bzl.

Done.

kchodorow:
Jenkins, test this please.

kchodorow (approved):

PR #48, opened 2017-08-03
Fix minor error in documentation.

petroseskinder:
Small fix in the documentation. Specifically, in the http archive, I include versioning information as well as a TODO statement. Both of which can and should be removed.


bazel-io:

petroseskinder:
I sign

PR #50, opened 2017-08-03
Add back in local range resolution

### LINKED ISSUE #47: Aether dependency resolution fails with "Invalid Range Result", generates partial workspace
As the title suggests, the generate_workspace tool after PR #45 ends up omitting some dependencies due to a problem with the new dependency resolution mechanism.

I noticed this problem when one of my dependencies attempted to pull in com.google.inject:guice:4.0 but the resulting generate_workspace.bzl did not contain corresponding maven_jar/java_library rules.

I've created a minimal test project: https://github.com/Zetten/bazel-generate-workspace-error

The pom.xml file describes Guava and Guice dependencies, but if the generate_workspace command is executed against it (may be run easily from the project dir with `bazel build :test_generate_workspace`) the resulting .bzl file does not contain the Guice dependency (the Guava deps are fine). The build log reports:

```
Aug 02, 2017 6:55:33 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for com.example:bazel-generate-error:0.0.0
Aug 02, 2017 6:55:34 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for com.google.guava:guava:22.0
Aug 02, 2017 6:55:34 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for com.google.code.findbugs:jsr305:2.0.2
Aug 02, 2017 6:55:34 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for com.google.errorprone:error_prone_annotations:2.0.18
Aug 02, 2017 6:55:35 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for com.google.j2objc:j2objc-annotations:1.1
Aug 02, 2017 6:55:35 PM com.google.devtools.build.workspace.maven.Resolver traverseDeps
INFO: 	Downloading pom for org.codehaus.mojo:animal-sniffer-annotations:1.14
Aug 02, 2017 6:55:35 PM com.google.devtools.build.workspace.maven.Resolver addDependency
WARNING: Could not resolve dependency com.google.inject:guice:4.0: Unable to find a version for com.google.inject:guice:4.0 due to Invalid Range Result
```

kchodorow:
If an artifact wants a version and it's resolvable, just use
that version. Don't make an extra round trip to fetch versions
unless the requested version doesn't work out.

Fixes #47.

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Aether.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Aether.java
@@ -69,7 +69,11 @@ public class Aether {
   /** Given an artifacts requests a version range for it. */
   List<String> requestVersionRange(Artifact artifact) throws VersionRangeResolutionException {
     VersionRangeRequest rangeRequest = new VersionRangeRequest(artifact, remoteRepositories, null);
-    VersionRangeResult result = repositorySystem.resolveVersionRange(repositorySystemSession, rangeRequest);
+    VersionRangeResult result = repositorySystem.resolveVersionRange(
+        repositorySystemSession, rangeRequest);
+    if (!result.getExceptions().isEmpty()) {
+      throw new VersionRangeResolutionException(result);
+    }
     return result.getVersions().stream().map(Version::toString).collect(toList());
   }
 
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
@@ -18,6 +18,10 @@ import static com.google.devtools.build.workspace.maven.ArtifactBuilder.InvalidA
 
 import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
+import javax.annotation.Nullable;
+import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
+import org.apache.maven.artifact.versioning.Restriction;
+import org.apache.maven.artifact.versioning.VersionRange;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.resolution.VersionRangeResolutionException;
 
@@ -45,7 +49,6 @@ class VersionResolver {
    */
   String resolveVersion(String groupId, String artifactId, String versionSpec)
       throws InvalidArtifactCoordinateException {
-
     List<String> versions;
     try {
       versions = requestVersionList(groupId, artifactId, versionSpec);
@@ -122,5 +125,40 @@ class VersionResolver {
    */
   public static VersionResolver defaultResolver() {
     return new VersionResolver(Aether.defaultOption());
+  }
+
+  /**
+   * Takes a version specification (as defined in
+   * http://maven.apache.org/enforcer/enforcer-rules/versionRanges.html) and finds a valid version
+   * that is likely to exist.  Basically: 1.2.3 is 1.2.3+, [1.2.3] is exactly 1.2.3, and then
+   * there is comma-separated range notation.
+   */
+  @Nullable
+  static String resolveVersionLocally(String unparsedVersion) {
+    VersionRange versionRange;
+    try {
+      versionRange = VersionRange.createFromVersionSpec(unparsedVersion);
+    } catch (InvalidVersionSpecificationException e) {
+      return null;
+    }
+    if (versionRange.getRecommendedVersion() != null) {
+      return versionRange.getRecommendedVersion().toString();
+    }
+
+    // There is a range or set of possible versions.
+    for (Restriction restriction : versionRange.getRestrictions()) {
+      // Look for an exact match.
+      if (restriction.getLowerBound().equals(restriction.getUpperBound())) {
+        return restriction.getLowerBound().toString();
+      }
+      // If this is a more complex version restriction, look for an inclusive bound.
+      if (restriction.isUpperBoundInclusive()) {
+        return restriction.getUpperBound().toString();
+      } else if (restriction.isLowerBoundInclusive()) {
+        return restriction.getLowerBound().toString();
+      }
+      // All bounds were exclusive.
+    }
+    return null;
   }
 }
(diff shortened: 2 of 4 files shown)


generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Aether.java
```diff
@@ -69,7 +69,11 @@ private Aether(RepositorySystem repositorySystem,
   /** Given an artifacts requests a version range for it. */
   List<String> requestVersionRange(Artifact artifact) throws VersionRangeResolutionException {
     VersionRangeRequest rangeRequest = new VersionRangeRequest(artifact, remoteRepositories, null);
-    VersionRangeResult result = repositorySystem.resolveVersionRange(repositorySystemSession, rangeRequest);
+    VersionRangeResult result = repositorySystem.resolveVersionRange(
+        repositorySystemSession, rangeRequest);
+    if (!result.getExceptions().isEmpty()) {
```
petroseskinder, line 74: thank you.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
```diff
@@ -123,4 +126,39 @@ private static String makeVersionRange(String versionSpec) {
...
+      // Look for an exact match.
+      if (restriction.getLowerBound().equals(restriction.getUpperBound())) {
+        return restriction.getLowerBound().toString();
+      }
+      // If this is a more complex version restriction, look for an inclusive bound.
+      if (restriction.isUpperBoundInclusive()) {
+        return restriction.getUpperBound().toString();
+      } else if (restriction.isLowerBoundInclusive()) {
+        return restriction.getLowerBound().toString();
+      }
```
petroseskinder, line 159: The above logic where you look for an inclusive bound introduces two different heuristics for selecting version ranges. One where we look for the highest version, and another where you look for an inclusive bound. _I am not a fan of this._ 

For illustration, suppose `something:something` has versions {1.0, 2.0, 3.0, 3.5, 4.0, 5.0}.  Let's say we make the following requests:
```
something:something:[3.0,)
something:something:[3.0, 5.5),
something:something:(2.9,)
something:something:[3.0, 5.0],
```

These should all evaluate to version 5.0. However, with the above logic, the first two will evaluate to 3.0, and the last two will evaluate to 5.0. 

These examples are not tested in the unit tests, and is what I was moaning and groaning about before #45.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/VersionResolver.java
```diff
@@ -123,4 +126,39 @@ private static String makeVersionRange(String versionSpec) {
...
+  static String resolveVersionLocally(String unparsedVersion) {
+    VersionRange versionRange;
+    try {
+      versionRange = VersionRange.createFromVersionSpec(unparsedVersion);
+    } catch (InvalidVersionSpecificationException e) {
+      return null;
+    }
+    if (versionRange.getRecommendedVersion() != null) {
+      return versionRange.getRecommendedVersion().toString();
+    }
```
petroseskinder, line 146: Can you separate out the above lines to a separate function, and remove the org.apache.maven.artifact.versioning classes?

When I was initially reading the code a month or two ago, this portion was very confusing to me. Especially the `getRecommendedVersion` portion, and the versionRange restrictions in 149. 

I believe the logic should be 
```java
static String resolveVersionLocally(String versionSpecification) {
    if ( isPin(versionSpecification) ) {
         return getPinnedVersion(versionSpecification);
    }
    return null;
}
```
For me at least, the org.apache.maven.artifact.versioning packaging introduces a lot of unnecessary mental overhead.

petroseskinder (commented):
Thanks for tackling the issue. There were three major requests I hammered on in the comments. 

First, if we use this guess based approach, I request that we locally resolve versions for _only pins_. No version ranges. The logic you have now, i.e. looking for an inclusive bound, introduces two different heuristics for selecting version ranges. One where we look for the highest version, and another where you look for an inclusive bound. _I am not a fan of this._  The third comment has a few examples where this is very problematic.

Second, you comment that we are taking 

> an extra round trip to fetch versions unless the requested version doesn't work out.

However, we are currently making only _one_ request per version specification. 

* For each version specification, we request all valid versions in the remote repository. This is our _only_ interaction with aether. 
* Soft pins like "3.0" are treated like unbounded version range e.g. "[3.0, )" 
* From this list of valid versions we select the appropriate one.
     - For soft pins, we simply get the oldest version. If the soft pin is valid, then oldest will be itself. If it doesn't we get the next oldest. This is the protocol we previously agreed upon.
     - For any other version range, we just get the highest version, which we agreed to in the past.

In fact, this change actually introduces additional requests. If our guessed version doesn't exist, then we will make two requests to aether. With the existing approach, we make exactly one request to aether per version specification. 

Third, the issue in #47 was due to version 4.0 not being listed in the [repository metadata](https://repo1.maven.org/maven2/com/google/inject/guice/maven-metadata.xml). That's an issue with the repository, _not us_. Your current solution resolves this specific issue.

However, in the process, it reintroduces a number of other issues mitigated in #45.

petroseskinder:
One additional clarification:

> Don't make an extra round trip to fetch versions unless the requested version doesn't work out.

We are currently making only _one_ request per version specification. 

* For each version specification, we request all valid versions in the remote repository. This is our _only_ interaction with aether. 
* Soft pins like "3.0" are treated like unbounded version range e.g. "[3.0, )" 
* From this list of valid versions we select the appropriate one.
     - For soft pins, we simply get the oldest version. If the soft pin is valid, then oldest will be itself. If it doesn't we get the next oldest. This is the protocol we previously agreed upon.
     - For any other version range, we just get the highest version, which we agreed to in the past.

All in all, the issue in #47 was due to version 4.0 not being listed in the [repository metadata](https://repo1.maven.org/maven2/com/google/inject/guice/maven-metadata.xml). That's an issue with the repository, not us. Your current solution resolves this specific issue, but it's important to acknowledge that this is a hack. We should minimize our reliance on this hack.

petroseskinder (changes requested):
Just noticed that "request changes" was an option.

petroseskinder:
@kchodorow bump.

damienmg:
Abandoning this PR since @kchodorow left the team. Feel free to take over.

PR #60, opened 2017-09-26
Add support for maven scopes.

### LINKED ISSUE #3: Support for Test and Runtime Maven Scopes
Within the Resolver class (com.google.devtools.build.workspace.maven package), the  [resolveEffectiveModel](https://github.com/bazelbuild/migration-tooling/blob/master/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java#L147) function simply skips over any dependencies that are not scoped as COMPILE. [Maven](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope) supports five other types of dependencies, of particular note are the test and runtime scope. It would be useful to add support to these scopes as well. 

Here is the relevant portion of the Resolver 
```java
public Model resolveEffectiveModel(ModelSource modelSource, Set<String> exclusions, Rule parent) {
    Model model = modelResolver.getEffectiveModel(modelSource);
    ...
    for (Dependency dependency : model.getDependencies()) {
      if (!dependency.getScope().equals(COMPILE_SCOPE)) {
        continue;
      }
      ...
    }
    ...
}

```

jkinkead:
This will generate targets for direct compile, test, runtime, and provided
dependencies, and will generate targets for transitive compile and runtime
dependencies.

Fixes #3 and #32 .

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
@@ -49,6 +49,7 @@ import org.apache.maven.model.io.DefaultModelReader;
 import org.apache.maven.model.locator.DefaultModelLocator;
 import org.apache.maven.model.resolution.UnresolvableModelException;
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.util.artifact.JavaScopes;
 
 /**
  * Resolves Maven dependencies.
@@ -66,8 +67,6 @@ public class Resolver {
   private static String unversionedCoordinate(Exclusion exclusion) {
     return exclusion.getGroupId() + ":" + exclusion.getArtifactId();
   }
-  
-  private static final String COMPILE_SCOPE = "compile";
 
   private final DefaultModelResolver modelResolver;
 
@@ -102,8 +101,9 @@ public class Resolver {
    * Given a local path to a Maven project, this attempts to find the transitive dependencies of
    * the project.
    * @param projectPath The path to search for Maven projects.
+   * @param scopes The scopes to look up dependencies in.
    */
-  public String resolvePomDependencies(String projectPath) {
+  public String resolvePomDependencies(String projectPath, Set<String> scopes) {
     DefaultModelProcessor processor = new DefaultModelProcessor();
     processor.setModelLocator(new DefaultModelLocator());
     processor.setModelReader(new DefaultModelReader());
@@ -114,7 +114,7 @@ public class Resolver {
     // Next, fully resolve the models.
     Model model = modelResolver.getEffectiveModel(pomSource);
     if (model != null) {
-      traverseDeps(model, Sets.newHashSet(), null);
+      traverseDeps(model, scopes, Sets.newHashSet(), null);
     }
     return pom.getAbsolutePath();
   }
@@ -137,7 +137,7 @@ public class Resolver {
     deps.put(rule.name(), rule);  // add the artifact rule to the workspace
     Model model = modelResolver.getEffectiveModel(modelSource);
     if (model != null) {
-      traverseDeps(model, Sets.newHashSet(), rule);
+      traverseDeps(model, Sets.newHashSet(JavaScopes.COMPILE), Sets.newHashSet(), rule);
     }
   }
   
@@ -146,7 +146,7 @@ public class Resolver {
    * file.
    */
   @VisibleForTesting
-  void traverseDeps(Model model, Set<String> exclusions, Rule parent) {
+  void traverseDeps(Model model, Set<String> scopes, Set<String> exclusions, Rule parent) {
     logger.info("\tDownloading pom for " + model.getGroupId() + ":"
         + model.getArtifactId() + ":" + model.getVersion());
     for (Repository repo : model.getRepositories()) {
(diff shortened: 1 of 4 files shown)


bazel-io:

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
```diff
@@ -146,7 +146,7 @@ public void resolveArtifact(String artifactCoord) {
    * file.
    */
   @VisibleForTesting
-  void traverseDeps(Model model, Set<String> exclusions, Rule parent) {
+  void traverseDeps(Model model, Set<String> scopes, Set<String> exclusions, Rule parent) {
```
petroseskinder, line 157: The `scopes` you pass into this function are the ones that you are considering worth traversing. Correct?

I see above, you are only traversing `compile` deps

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
```diff
@@ -163,18 +163,34 @@ void traverseDeps(Model model, Set<String> exclusions, Rule parent) {
...
     // DependencyManagement dependencies don't have scope.
-    if (dependency.getScope() != null && !dependency.getScope().equals(COMPILE_SCOPE)) {
-      return;
+    if (scope != null) {
+      if (parent == null) {
+        // Top-level scopes get pulled in based on the user-provided scopes.
+        if (!topLevelScopes.contains(scope)) {
+          return;
+        }
+      } else {
```
petroseskinder, line 192: Can you expand on what you mean by "relabel the transitive dependencies"?

For dependencies that are not scoped as compile or runtime, what would they be relabeled as?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
jkinkead: These scopes filter which dependencies we'll include from the root project. At least, that's the intent - I believe the `parent` parameter indicates the parent *dependency*, not the parent *project*, yes?

Another option would be to have a filter in [the `resolvePomDependencies` function](https://github.com/bazelbuild/migration-tooling/blob/cc88b8da996c222a3819aa006e795b68dbe983bd/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java#L115-L118), before calling `traverseDeps` for the first time.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
jkinkead: The issue is that transitive dependencies are handled identically regardless of the scope on which the root project depends on them.

For example, if the root project depends on `junit` in the `test` scope, all of `junit`s `compile` dependencies will be treated as a `compile` dependency of the root project, instead of as a `test` dependency.

This doesn't matter much if we're not generating a `java_test` target.

petroseskinder: I see. So you are saying, the scope of the parent should propagate to all of its children. That makes sense. 

Minor nit, but can you rephrase the TODO so that it says something like:
```
// TODO (bazel-devel): Relabel the scope of transitive dependencies so they match the scope of their parent.
 ```
From a cursory glance, this does not cause issues, if we hold off on making this change. However, if you are able to, can you try incorporating them into this pull request?

File: generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java (hunk 1 above)
Comment (petroseskinder) [reply generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java:157]: >  I believe the parent parameter indicates the parent dependency, not the parent project, yes?

That is correct. It is the parent dependency. 
> Another option would be to have a filter in the resolvePomDependencies function, before calling traverseDeps for the first time.

I am not confident that is a better approach, at least if we do it manually. Aether has a rich set of functions for filtering and modifying the universe of dependencies. However, that is not completely integrated. This seems like a happy medium.

File: generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java [hunk 3]
```diff
@@ -137,7 +137,7 @@ public void resolveArtifact(String artifactCoord) {
     deps.put(rule.name(), rule);  // add the artifact rule to the workspace
     Model model = modelResolver.getEffectiveModel(modelSource);
     if (model != null) {
-      traverseDeps(model, Sets.newHashSet(), rule);
+      traverseDeps(model, Sets.newHashSet(JavaScopes.COMPILE), Sets.newHashSet(), rule);
```
petroseskinder: Small nit. Can you extract the set of valid scopes into a global variable? For example:

` private static final Set<String> VALID_SCOPES = Sets.newHashSet(JavaScopes.COMPILE);`

Slightly more readable.

petroseskinder (changes requested):
Thanks again for the PR. Commented on a few nits.
@cgrushko @damienmg

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
jkinkead: It should propagate as in the maven rules, linked in the comment. These match except for compile->runtime dependencies, which is handled as runtime instead of compile.

Are you asking just to update the comment, or to update the logic fully? I can't do the latter without adding a scope field to `Rule`, and I wasn't sure if that was desirable.

petroseskinder: Just the change to the comments.

jkinkead: Done.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
jkinkead: Done - although this line should actually use an empty set, which I updated.

jkinkead (commented):
Comments addressed.

petroseskinder:
Thank you. LGTM @cgrushko @damienmg

damienmg (approved):

damienmg:
jenkins: test this please

PR #65, opened 2017-09-27
Implement ModelResolver.addRepository(Repository, boolean) correctly.

jkinkead:
This correctly honors the `replace` parameter. It was ignored before, which was making it effectively be treated as `false`.

Per [the reference implementation](https://github.com/apache/maven/blob/542a7a89156263b34d1472e9d9c1a2795afccd2d/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/DefaultModelResolver.java#L113-L144), the single-argument `addRepository` should not replace (although it wasn't before).

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
@@ -17,7 +17,9 @@ package com.google.devtools.build.workspace.maven;
 import static com.google.devtools.build.workspace.maven.Rule.MAVEN_CENTRAL_URL;
 import static java.util.stream.Collectors.toList;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import java.io.IOException;
@@ -58,15 +60,19 @@ public class DefaultModelResolver implements ModelResolver {
 
   private final static Logger logger = Logger.getLogger(
       MethodHandles.lookup().lookupClass().getName());
-  private static final Repository MAVEN_CENTRAL = new Repository();
 
+  private static final Map<String, Repository> DEFAULT_REPOSITORIES;
   static {
-    MAVEN_CENTRAL.setId("central");
-    MAVEN_CENTRAL.setName("default");
-    MAVEN_CENTRAL.setUrl(MAVEN_CENTRAL_URL);
+    Repository mavenCentral = new Repository();
+    mavenCentral.setId("central");
+    mavenCentral.setName("default");
+    mavenCentral.setUrl(MAVEN_CENTRAL_URL);
+
+    DEFAULT_REPOSITORIES = ImmutableMap.of(mavenCentral.getId(), mavenCentral);
   }
 
-  private final Set<Repository> repositories;
+  @VisibleForTesting
+  final Map<String, Repository> repositoriesById;
   private final Map<String, ModelSource> ruleNameToModelSource;
   private final DefaultModelBuilder modelBuilder;
   private final Aether aether;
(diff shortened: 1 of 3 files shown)


bazel-io:

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
```diff
@@ -182,12 +182,19 @@ public ModelSource resolveModel(Parent parent) throws UnresolvableModelException
...
   // don't add @Override.
   public void addRepository(Repository repository) {
-    repositories.add(repository);
+    addRepository(repository, false);
   }
 
   @Override
   public void addRepository(Repository repository, boolean replace) {
-    addRepository(repository);
+    if (repositories.contains(repository)) {
```
petroseskinder: Thank you. The `addRepositories(Repository repository, boolean replace)` is never called, so we skimped out on including the replace functionality. 

Assuming it's not particularly annoying, can you add a simple test case?

There isn't any `DefaultModelResolverTest` so you would need to create one, and mock the repositories.

jkinkead: Done.

I also updated the repository store, since the previous implementation was relying on the internal implementation of [Repository.equals](https://maven.apache.org/ref/3.5.0/maven-model/apidocs/org/apache/maven/model/RepositoryBase.html#equals(java.lang.Object)), which is not documented as operating only on ID (even though it appears to be).

petroseskinder: Good catch on the equals.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
```diff
@@ -187,19 +195,14 @@ public void addRepository(Repository repository) {
...
   @Override
   public void addRepository(Repository repository, boolean replace) {
-    if (repositories.contains(repository)) {
-      if (replace) {
-        repositories.remove(repository);
-        repositories.add(repository);
-      }
-    } else {
-      repositories.add(repository);
+    if (replace || !repositoriesById.containsKey(repository.getId())) {
```
petroseskinder: nit: can you add a simple comment for the boolean logic?
Although the new version is more concise, it took me a
second to remember the conditions.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
```diff
@@ -58,15 +60,19 @@
...
   static {
-    MAVEN_CENTRAL.setId("central");
-    MAVEN_CENTRAL.setName("default");
-    MAVEN_CENTRAL.setUrl(MAVEN_CENTRAL_URL);
+    Repository mavenCentral = new Repository();
+    mavenCentral.setId("central");
+    mavenCentral.setName("default");
+    mavenCentral.setUrl(MAVEN_CENTRAL_URL);
+
+    DEFAULT_REPOSITORIES = ImmutableMap.of(mavenCentral.getId(), mavenCentral);
```
petroseskinder, line 71: 👍 to this change

generate_workspace/src/test/java/com/google/devtools/build/workspace/maven/DefaultModelResolverTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Tests for {@link DefaultModelResolver}.
+ */
+@RunWith(JUnit4.class)
+public class DefaultModelResolverTest {
+  /** Test that addRepository doesn't replace existing repositories by default. */
+  @Test
+  public void testAddRepository_defaultReplace() {
+    Repository repoA1 = new Repository();
```
petroseskinder, line 18: First thank you for the tests. One nit

Since each test is effectively doing the same thing, can you create the following utility functions

```java
private Repository createRepository(String id, String url) {
     Repository repo = new Repository();
     repo.setId(id);
     repo.setUrl(url);
}

private void assertThatContainsRepo(
      DefaultModelResolver resolver, Repository repo) {...} 

private void assertThatUrlNotChanged(
       DefaultModelResolver resolver, Repository repo, String expectedUrl) {...}
```

petroseskinder (changes requested):
Thank you. It looks good overall, barring a few nits.

PR #66, opened 2017-09-27
Handle file-based repositories.

### LINKED ISSUE #64: File repositories don't work
A `repository` pointing at the filesystem fails to resolve any dependencies, due to some assumptions made about POM file locations.

I have a fix for this (although it only works for absolute-path repositories per issue #63).

jkinkead:
This adds support for repositories in the local file system.

Fixes #64 .

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
@@ -26,6 +26,7 @@ import java.io.InputStreamReader;
 import java.lang.invoke.MethodHandles;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.charset.Charset;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


bazel-io:

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/Resolver.java
```diff
@@ -320,8 +321,11 @@ static String getSha1Url(String url, String extension) {
   private String downloadSha1(Rule rule) {
     String sha1Url = getSha1Url(rule.getUrl(), rule.getArtifact().getExtension());
     try {
-      HttpURLConnection connection = (HttpURLConnection) new URL(sha1Url).openConnection();
-      connection.setInstanceFollowRedirects(true);
+      URLConnection connection = new URL(sha1Url).openConnection();
+      if (connection instanceof HttpURLConnection) {
```
petroseskinder, line 325: Thank you again. 

In the situation that it is not an `HttpURLConnection`, it would be a [`JarURLConnection`](https://docs.oracle.com/javase/7/docs/api/java/net/JarURLConnection.html). I doubt this should occur, so can you throw an exception? 

Additionally, can you separate the sha functionality into a separate class, and add a simple test case for when this URLConnection fails? 

`Resolver.java` has gotten fairly bloated, and I am trying to steer it so that `Resolver.java` has less functionality backed in. (see `VersionResolver.java` as an example)

jkinkead: The connection can be an instance of any type that the [`URLStreamHandler`](https://docs.oracle.com/javase/8/docs/api/index.html?java/net/URLStreamHandler.html) for the protocol returns. For a file, this appears to be `sun.net.www.protocol.file.FileURLConnection` (at least on the JRE I'm using).

Both the sha and pom files will have the protocol `file` if the repository has a `file` URL.

Issue #14, opened 2017-06-13
Add support for artifactory and private repositories


petroseskinder:
# Migrating from [issue 498](https://github.com/bazelbuild/bazel/issues/498) in main repository. 
This is already implemented on the Bazel side, so similar code can probably be used for generate_workspace. It should use all of the local (settings.xml) configuration options.

## An example from the original post:

I think I've run into a similar requirement; I'm trying to use robolectric as in:

`bazel run //src/tools/generate_workspace -- --artifact=org.robolectric:robolectric:3.1.2`

However, robolectric depends on `maven-ant-tasks` which depends on some `org.codehaus.plexus` artifacts. Codehaus.org no longer appears to host code, but bazel is attempting to use it to retrieve the plexus assets:

```
ERROR: Unable to resolve Maven model from http://snapshots.maven.codehaus.org/maven2/org/codehaus/plexus/plexus-utils/1.0.4/plexus-utils-1.0.4.pom: 1 problem was encountered while building the effective model
[FATAL] Non-parseable POM http://snapshots.maven.codehaus.org/maven2/org/codehaus/plexus/plexus-utils/1.0.4/plexus-utils-1.0.4.pom: Expected root element 'project' but found 'html' (position: START_TAG seen \n... @2:7) @ line 2, column 7
```

I think that if I could specify a settings.xml, then I might be able to resolve this with http://www.codehaus.org/mechanics/maven/

or:

http://stackoverflow.com/questions/30288097/how-to-resolve-codehaus-service-unavailable-maven-build-error

But since this issue seems to state that there is no way to specify a `settings.xml` when using `generate_workspace`, does anyone know of some workaround I can use in the meantime?

Thanks!

epollan:
I'm blocked by this too, I believe.  Seems like configuring Aether to read settings.xml would do the trick, no?  From https://wiki.eclipse.org/Aether/Creating_a_Repository_System_Session:

> If you seek a closer cooperation with Apache Maven and want to read configuration from the user's settings.xml, you should have a look at the library org.apache.maven:maven-settings-builder which provides the necessary bits. The method AntRepoSys.getSettings() from the Aether Ant Tasks can serve as inspiration for your own code. But please direct any questions regarding usage of that library to the Maven mailing list.

petroseskinder:
Yes, in fact, it would.

epollan:
OK.  I'll try to put together a PR.

petroseskinder:
[...]

PR #71, opened 2017-10-24
Enable custom repositories.

mweiden:
Enables the specification of custom repositories in the `transitive_maven_jar` function.

For more, see https://github.com/bazelbuild/migration-tooling/issues/14#issuecomment-338764262

This is my first PR to a bazel project, so this will likely need some code review and guidance to reach master. Thanks in advance!

### DIFF
--- a/generate_workspace/src/main/java/com/google/devtools/build/workspace/GenerateWorkspace.java
+++ b/generate_workspace/src/main/java/com/google/devtools/build/workspace/GenerateWorkspace.java
@@ -30,13 +30,11 @@ import java.util.List;
 import java.util.Set;
 import java.util.logging.Logger;
 
-/**
- * Generates a WORKSPACE file for Bazel from other types of dependency trackers.
- */
+/** Generates a WORKSPACE file for Bazel from other types of dependency trackers. */
 public class GenerateWorkspace {
 
-  private final static Logger logger = Logger.getLogger(
-      MethodHandles.lookup().lookupClass().getName());
+  private static final Logger logger =
+      Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
 
   private final Resolver resolver;
   private final List<String> inputs;
@@ -58,8 +56,13 @@ public class GenerateWorkspace {
     }
 
     try {
-      GenerateWorkspace workspaceFileGenerator = new GenerateWorkspace(
-          args, options.outputDir, options.directToWorkspace, options.aliases);
+      GenerateWorkspace workspaceFileGenerator =
+          new GenerateWorkspace(
+              args,
+              options.outputDir,
+              options.directToWorkspace,
+              options.repositories,
+              options.aliases);
       workspaceFileGenerator.generateFromPom(options.mavenProjects, options.scopes);
       workspaceFileGenerator.generateFromArtifacts(options.artifacts);
       workspaceFileGenerator.writeResults();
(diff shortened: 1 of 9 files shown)


bazel-io:

mweiden:
@google I signed it!

wstrange:
@mweiden Nice.  

Dumb question: Can this use ~/.m2/settings.xml to authenticate against a private repo such as Artifactory?  Is there a way to specify the settings.xml location?

petroseskinder (commented):
Thank you! The code itself generally looks good. What are you using to test? Are you hosting a maven-central mirror, internally with Artifactory? In the ideal situation, this would also include some unit tests. However, I don't know off the top of my head how difficult/tedious that would be.

petroseskinder:
> Can this use ~/.m2/settings.xml to authenticate against a private repo such as Artifactory? Is there a way to specify the settings.xml location?

@wstrange. As it stands, this PR allows one to pass urls to maven mirrors as arguments to `generate_workspace`. Perhaps, we leave settings.xml functionality (#7) to a separate PR.

mweiden:
@petroseskinder I don't have tests for this. You're right, they should be added. I will try to work on this on Friday.

mweiden:
@petroseskinder tests added. 👀 please!

generate_workspace/src/main/java/com/google/devtools/build/workspace/GenerateWorkspace.java
```diff
@@ -69,13 +79,22 @@ public static void main(String[] args) throws InterruptedException {
...
+  private GenerateWorkspace(
+    String[] args,
+    String outputDirStr,
+    boolean directToWorkspace,
+    List<String> repositories,
+    List<Rule> aliases
+  ) throws IOException {
+    DefaultModelResolver defaultModelResolver = new DefaultModelResolver();
+    for (String repo : repositories) {
+      defaultModelResolver.addUserRepository(repo);
```
petroseskinder: Perhaps, overloading the constructor? Admittedly, the base constructor has four arguments as is, so I can see why you would explicitly add the repositories like this.

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
```diff
@@ -187,6 +189,10 @@ public ModelResolver newCopy() {
     return new DefaultModelResolver(aether, repositories, ruleNameToModelSource, modelBuilder);
   }
 
+  public VersionResolver getVersionResolver() {
+    return this.versionResolver;
+  }
```
petroseskinder: This is more of a pet peeve. Any way to use the default VersionResolver?

generate_workspace/src/main/java/com/google/devtools/build/workspace/GenerateWorkspaceOptions.java
```diff
@@ -53,6 +53,12 @@
   public List<String> mavenProjects = new ArrayList<>();
 
   @Parameter(
+          names = {"--repository", "-r"},
+          description = "Repository to query for artifacts."
```
petroseskinder: Nit. Can you replace `repository` with `repositories` or `repository_list`?
And can you replace the description so that it is
```
List of remote maven repositories to query for artifacts
```

petroseskinder (commented):
Thank you @mweiden! 

It generally looks good. 

There were a few nits. These are mostly things like indentation, import statements and the like. I think if you ran the [Google Java Formatter](https://github.com/google/google-java-format) on your code it would take care of most of them. Can you run the formatter?

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
mweiden: I've made the other changes. I think this problem is reflective of a larger issue in the code base: all of these classes are very tightly coupled. In this case, the `DefaultModelResolver` has a `VersionResolver` which depends on an instance of `Aether`. The `Resolver` class needs an instance of `DefaultModelResolver`, but also requires a `VersionResolver` which in turn needs an instance of `Aether` that needs to see the same `RemoteRepositories` as the `DefaultModelResolver`. I believe this level of refactor is outside the scope of this pull request.

mweiden:
@petroseskinder I've made all but one of the changes you requested, run the code formatter, and updated the jar. Please see my comments regarding the default version resolver in the last outdated diff. I hope that all this looks good and is ready to merge.

petroseskinder (approved):

petroseskinder:
Test this please

generate_workspace/src/main/java/com/google/devtools/build/workspace/maven/DefaultModelResolver.java
petroseskinder: I am in agreement with you about the pervasive tight coupling. One precaution before investing considerable effort on that front. The bazel team is moving towards making [bazel-deps](https://github.com/johnynek/bazel-deps) the tool of choice for managing third party maven dependencies, so it may be a better use of your time to get accustomed to that tool.

mweiden: @petroseskinder OK, ty for the input.

PR #74, opened 2017-11-20
Fixes multiple repository issue.

sjincho:
The Repository instances are identified by id field, thus creating a set
of Repository instances w/o id (id = null) will cause the set to contain
only the first instance.


bazel-io:

PR #87, opened 2018-05-16
Avoid dyndns.org lookup

yinanfang:
dyndns.org is broken and doesn't give any response. We have a couple hundred dependencies and this the url connection won't stop until it times out after like 30 seconds. This results in that the entire dependency generation process takes longer than a couple hours.

We should avoid trying to download pom file from their site. Adding a simple blacklist can fix this issue.


yinanfang:
I signed it!

PR #97, opened 2019-01-30
Create CODEOWNERS

dslomov:
Adding global CODEOWNER


dslomov:
Alternatively we could make this repo readonly (since it is deprecated)

petroseskinder:
@dslomov I would personally prefer we archive this repo. 
It was decided in Fall 2017 that bazel-deps would be the blessed approach. 
In that ensuing time, I have noticed many people valiantly trying to get this tool to work.

petroseskinder (changes requested):
Can we mark archive instead? I haven't done any active development on this project since around when Kristina left (~fall 2017).

